PR #32, opened 2015-01-10
Added SVG placeholder icons

MichaelArestad:
* includes changes to Gruntfile.js
* `grunt` now runs the Watch task (currently only watches SVG files)
* `grunt lint` runs linters


MichaelArestad:
If there's a preference on which folder font/icon/other assets should live, I'll move the folders there and update the PR.

Gruntfile.js
```diff
@@ -33,6 +85,7 @@ grunt.registerTask( "update-authors", function() {
 	});
 });
 
-grunt.registerTask( "default", [ "jshint", "jscs" ] );
+grunt.registerTask( "default", [ "watch" ] );
```
arschmitz: I would not have watch run on the default task I never ever use watch. Im all for having a watch task just not as the default.

Gruntfile.js
```diff
@@ -13,6 +16,55 @@ grunt.initConfig({
...
+				ext: ".svg"
+			} ]
+		}
+	},
+
+	// Combines SVGs into single file
+	svgstore: {
+		defaults: {
+			options: {
+				prefix: "icon-", // This will prefix each ID
```
arschmitz: We previously decided to follow the jQuery js style guide unless we come up with a strong reason not to http://contribute.jquery.org/style-guide/js/

So comments should be on the line above with a blank line preceding them here and line below

Gruntfile.js
MichaelArestad: I can remove it making it `grunt watch` instead, but it's pretty handy when working with Sass/LiveReload (when we get that set up). `

arschmitz: Yeah i'm all for having a watch task

Gruntfile.js
MichaelArestad: Rad. Changing.

arschmitz:
Also can I get you to sign the CLA at http://contribute.jquery.org/CLA/ Please :-)

MichaelArestad:
Done. :)

arschmitz:
A personal preference thing here and ill let other weigh in but i don't like to see dist or min files in repos. I would make a build task and have that run on the default task. Cuts down own bloat in the repo and also avoids issues with issues things being updated for fixed but the update to dist files getting missed. ( iv seen this a lot with minor fixes like typos )

MichaelArestad:
Works for me. (same preference) Removing the min/build files.

sfrisk:
Does anyone have any other concerns about this PR?

arschmitz:
Looks good to me :+1:

MichaelArestad:
Works for me. I can add fallback bits in another PR.

arschmitz:
@MichaelArestad oh sorry just went to land this, and realized you are using your s/n name, instead of your real name, in your git config. Think you could fix that? ( I know its just a space but it needs to match to pass the checks )

Issue #29, opened 2014-12-24
File structure


MichaelArestad:
I suggest an atomic structure with small self-contained modules. I prefer a version inspired by a Dale Sande [article](http://gist.io/4436524) [(or video)](https://www.youtube.com/watch?v=EmNcD3b3ZtI).

Here's how I propose each module is structured:

```SCSS
// Module structure
// 1. main file with styles
// 2. functions used with main file
// 3. mixins used with main file
// 4. variables used with main file

|- typography/
|--- functions.scss  // 2
|--- mixins.scss     // 3
|--- typography.scss // 1
|--- variables.scss  // 4
```

And how the overall file structure might look:

```SCSS
// 1. _utilities contains global mixins module, normalize, etc
// 2. atoms contains modules like buttons
// 3. molecules contains  modules like nav_horizontal
// 4. organisms contains more complex modules like banner
// 5. views contains individual page/view templates like admin

|- sass/
|--- _utilities/ // 1
|--- atoms/      // 2
|--- molecules/  // 3
|--- organisms/  // 4
|--- views/      // 5
|--- style.scss
```

### Pros
* This method keeps everything pretty tidy. The variables, functions, mixins, extends (dependencies) relevant to a module together rather than scattering them around via a massive variables file, etc.
* The module name is the same as the `.base` name of the module's rules and components. If the module was `calendar`, it's rules might be `calendar_day.is-today` (I'm referencing the BEM naming convention I referenced [here](https://github.com/jquery/css-chassis/issues/4#issuecomment-68012541). This makes it easy to avoid naming conflicts as it's easy to spot the base elements in use. There is an added benefit of knowing where to find which file a class was declared in by just knowing the class name.
* There are no mystery junk drawer files. For example, there is no giant universal file with a series of placeholder selectors and no way to know which selectors exist and where they are used.
* Because of it's structure, style.css ends up being a pretty rad little table of contents that might look like this:
```SCSS
@import "atoms/colors/colors",  // Global color variables
	"_utilities/mixins/mixins", // Global mixins
	"_utilities/clearings",

	"atoms/typography/typography",
	"atoms/media",
	"atoms/animations",
	"atoms/buttons",
	"atoms/icons/genericons",

	"molecules/nav-horizontal",
	"molecules/nav-off-screen",
	"molecules/alerts",

	"organisms/contact-form",
	"organisms/calendar",

	"views/main",
	"views/settings";
```

### Cons
* Looks weird and scary at first. It's not.
[...]

Issue #4, opened 2014-11-03
Naming convention


scottgonzalez:
BEM, SUIT, OOCSS, etc.?

People seemed pretty open to what naming we use, but there also seemed to be some consensus on just picking names that make sense and being flexible instead of being tied into a specific naming system.

battaglr:
Usually when discussing about naming conventions, we end up talking about architecture. Of course, the two are closely related and it's understandable. *Maybe we could open another issue to discuss specifically about the architecture.*

Anyhow, AFAIK —and **please** correct me if I'm wrong:

+ OOCSS doesn't have a canonical naming convention. Fortunately, as some have mention, @csswizardry has done an amazing job on this topic, but it's not that different from BEM's approach... and I like them both. But it's not an OOCSS naming convection rather a BEM variant –not like if that's a big deal.
+ SMACSS attempts to suggest a naming convention, but @snookca doesn't intent to be too prescriptive about it.
+ BEM has a more rigid naming convention, but encourages developers to adapt it to their needs.
+ SUITCSS by @necolas is the most extensively documented and maybe prescriptive naming convention so far and has a lot of great ideas –it's important to notice that it takes a lot from BEM and some from SMACSS.

IMHO we could define a new one if we feel like it, or just go with one of the previously mentioned. In my experience, a sane naming convention should consider the following:

+ Helper, utility or general purpose classes.
+ Components –blocks, or however we like to call them.
+ Sub-components –elements, or however we like to call them.
+ Modifiers.
+ States —i.e.: `.is-active`.
+ Behavior hooks —this would be how we would like to reference elements to be manipulated with JavaScript.

I know I haven't proposed any naming convention, but my intention was to take a little step back to see if we need to explicitly define what we need/want and build upon that.

Hope it helps.

arschmitz:
@battaglr Thanks for all the information and input. As you can see we have had a TON of discussion around this, ( what you see in this issue is only the tip of the iceberg ). 

We through all this discussion, came to the decision as a team, that we wanted to go with a BEM naming convention, for a variety of reasons. The core of these reasons came down to, performance, flexibility, and ease of use, for the cases we are trying to address. One thing I would like to mention, that I think is a common misconception, is that BEM  has a super strict naming convention. This was something I had been wrong about too, until we talked with @veged and others from Yandex ( creators of BEM ).

[...]

Issue #3, opened 2014-11-03
Browser support


scottgonzalez:
In the planning meeting we said:
- IE8+ (possibly 9+ or 10+ based on the release date of v1.0.0)
- Chrome (Current - 1)
- Firefox (Current - 1)
- Opera (Current - 1)
- Safari 5.1+
- iOS 6.1+
- Android 2.3, 4+

However, we should probably determine a list of officially supported devices that we'll test against as well.

PR #33, opened 2015-01-10
Initial Sass file structure

MichaelArestad:
To address https://github.com/jquery/css-chassis/issues/29.

* Added basic file structure with a few example modules
* Added Sass and Autoprefixer tasks to Gruntfile.js


package.json
```diff
@@ -1,48 +1,50 @@
...
+      "url": "https://github.com/jquery/css-chassis/blob/master/LICENSE.txt"
+    }
+  ],
+  "dependencies": {},
+  "devDependencies": {
+    "commitplease": "2.0.0",
+    "grunt": "0.4.2",
+    "grunt-autoprefixer": "^2.1.0",
+    "grunt-contrib-jshint": "0.10.0",
+    "grunt-contrib-sass": "^0.8.1",
```
arschmitz: We previously decided, to avoid an annoying ruby dependency when its likely most people have no other need for ruby, to use libsass. So this would be just `grunt-sass` for that.

Gruntfile.js
```diff
@@ -13,6 +16,52 @@ grunt.initConfig({
...
+				// Target-specific options go here.
+				// browser-specific info: https://github.com/ai/autoprefixer#browsers
+				// DEFAULT: browsers: [ "> 1%", "last 2 versions", "ff 17", "opera 12.1" ]
+				browsers: [
+				"> 1%",
+				"last 2 versions",
+				"ff 17",
+				"opera 12.1",
+				"ie 8",
+				"ie 9"
```
arschmitz: This could change but as of right now our support is
IE8+
Chrome (Current - 1)
Firefox (Current - 1)
Opera (Current - 1)
Safari 5.1+
iOS 6.1+
Android 2.3, 4+

https://github.com/jquery/css-chassis/issues/3

Gruntfile.js
```diff
@@ -13,6 +16,52 @@ grunt.initConfig({
...
+	},
+
+	autoprefixer: {
+		dist: {
+			options: {
+				map: true,
+
+				// Target-specific options go here.
+				// browser-specific info: https://github.com/ai/autoprefixer#browsers
+				// DEFAULT: browsers: [ "> 1%", "last 2 versions", "ff 17", "opera 12.1" ]
```
arschmitz: Should remove default comment

style.css
```diff
@@ -0,0 +1,3 @@
+.clearfix:before,.clearfix:after{content:"";display:table}.clearfix:after{clear:both}
+
+/*# sourceMappingURL=style.css.map */
```
arschmitz: Missing blank line at end of file

package.json
MichaelArestad: Oops. Fixing.

style.css
MichaelArestad: I'm not sure how to add a blank line as this file is generated (other than manually doing it each time). I don't see options for it in the `grunt-sass` task. Ideas?

Gruntfile.js
MichaelArestad: Removed.

Gruntfile.js
MichaelArestad: Updated to support those.

package.json
arschmitz: Added `grunt-sass` but forgot to remove `grunt-contrib-sass`

style.css
```diff
@@ -0,0 +1,6 @@
+.clearfix:before, .clearfix:after {
+  content: "";
+  display: table; }
+.clearfix:after {
+  clear: both; }
+/*# sourceMappingURL=style.css.map */
```
arschmitz: Hmm i'm not sure about how to get it to add the blank link ill have to look into it. Even github view complains about the lack of blank line so i'm sure there must be a way to get it to add it.

MichaelArestad: We could just not include it as it's a compiled file.

jzaefferer:
> We could just not include it as it's a compiled file.

I can't find discussion about this, but while dealing with sass files on an unrelated problem, I wanted to provide some input on that.

The project I'm working on uses sass files and checks in compiled CSS files in parallel. These are minified, so each time a scss file changes, there's a one-line diff on the compiled CSS file. This is fine for serial development on master, since anyone can pull and look at pages without first having to build something. This is painful when rebasing anything, since every commit that changes any scss file will produce a conflict on the CSS file.

Maybe there's a strategy to avoid those conflicts that I'm not aware of? Or is it better to force everyone to compile scss themselves?

Gruntfile.js
```diff
@@ -13,6 +16,53 @@ grunt.initConfig({
...
+				"last 2 versions",
+				"safari >= 5.1",
+				"ios >= 6.1",
+				"android 2.3",
+				"android >= 4",
+				"ie >= 8"
+				]
+			},
+			src: [
+			"*.css"
```
jzaefferer: Missing indent. Though for single items, you shouldn't need an array anyway, just pass the string directly.

Gruntfile.js
```diff
@@ -13,6 +16,53 @@ grunt.initConfig({
...
+			]
+		}
+	},
+	watch: {
+		sass: {
+			files: [ "scss/**/*.scss" ],
+			tasks: [ "sass", "autoprefixer" ],
```
jzaefferer: There should be an alias that can be called directly (when not using watch), that could be used here as well, like "build".

scss/_utilities/_clearfix.scss
```diff
@@ -0,0 +1,21 @@
+// ==========================================================================
+// Clearfix
+//
+// You can either use the mixin or @extend .clear.
```
jzaefferer: I've used neither syntax before, so this doesn't help me understand how to use this. And the `@extend .clear` looks odd, since only `.clearfix` is defined below.

Gruntfile.js
MichaelArestad: Good catch. Will fix.

scss/_utilities/_clearfix.scss
MichaelArestad: Yep. That's an old comment leftover. Will remove/clarify.

MichaelArestad:
> This is painful when rebasing anything, since every commit that changes any scss file will produce a conflict on the CSS file.

That's partially why I might prefer to just not include the compiled file. As we're eventually going to have a build system, there's not a huge need to have the compiled file at all.

jzaefferer:
I suppose there is no way around for contributors to compile scss to css, right? So might as well go with that.

Btw. I remember something about only using scss features that are in future CSS specs (or something like that). Does that apply to the few lines that are in this PR?

arschmitz:
@jzaefferer the discussion was brief and in #css-chassis i would say this is still open for discussion

Issues like this are part of why i would prefer to not see built files in the repo.

@sfrisk any thoughts here?

arschmitz:
@jzaefferer yeah actually having built files could add confusion as you might end up with beginner contributors submitting patches on the css files rather then the scss

@jzaefferer as far as future features that is what was discussed. Im not entirely sure on whats in the specs so far. If we are going to have a rule like this we should probably add this to the css style guide to make it clear whats ok not ok to use. @sfrisk thoughts? we will want this settled before we commit any scss i would think?

Gruntfile.js
MichaelArestad: Added.

sfrisk:
I'm leaning towards removing the .css build files and sticking to .css.  I definitely could see the issue with beginners submitting fixes to the css rather than the scss.

Maybe edit the readme as part of this PR (or a separate one) with a "Getting Started" section that explains how to get everything up and running?

arschmitz:
:+1:  for adding getting started to readme

jzaefferer:
That information should go into `CONTRIBUTING.md`, though there should be a link to it in the readme.

arschmitz:
@jzaefferer that sort of information is in the readme in all other repos though getting started is probably not the right section name as much as "building chassis" 

https://github.com/jquery/jquery-ui#building-jquery-ui
https://github.com/jquery/jquery-mobile#buildcustomization
https://github.com/jquery/jquery#how-to-build-your-own-jquery

jzaefferer:
So were [mostly](https://github.com/jquery/jquery-ui/blame/79c4fa1e7a09a3d28f0f8ff7a5b9f41a6d6e123d/README.md#L15) written [before](https://github.com/blog/1184-contributing-guidelines) contributing guidelines were a feature. They should be moved.

arschmitz:
This does not seem to keep with the spirit of contributing guidelines outlines in the blog post you link to. These should be a set of guidelines for contributors to follow when doing a patch. basically contribute.jquery.org. These are setup instructions the exactly what is whats meant to be in a readme http://en.wikipedia.org/wiki/README

sfrisk:
How about for now we just have a "Getting Started / Building Chassis" on the Readme for now, so all the necessary information someone might want to know after arriving to the repo is there on the first page. Might help with attracting contributors if they can find out the state of things and how to get started right there on the home page.

On that note, @arschmitz, for your #34 PR, maybe add to the readme the how-to for getting perf testing up and running?

MichaelArestad:
I'll remove the build files and add info to the README. 

Quick note: will miss the meeting today.

MichaelArestad:
Added a Getting Started section to the README and removed the built Sass file

Gruntfile.js
```diff
@@ -13,6 +16,51 @@ grunt.initConfig({
...
+		}
+	},
+
+	autoprefixer: {
+		dist: {
+			options: {
+				map: true,
+
+				// Target-specific options go here.
+				// browser-specific info: https://github.com/ai/autoprefixer#browsers
```
arschmitz: There comments seem more like documentation for autoprefixer then comments about what we are actually doing here.

MichaelArestad: I think I'll just remove the comments here.

arschmitz:
@sfrisk agreed. I was looking around at how other projects do this. I don't think we need to explain git. Maybe just add a link to https://help.github.com/articles/set-up-git/ for setup of third party things like git and NPM there are great resources we don't have to maintain. Looking at other projects that use preprocessors they add pretty much  no documentation even on building. I think since barrier to contribution, is a big deal to us, we want some documentation here though.

Personally here i would stick to the specific dependencies they need to install ( not how just list them with links ), and then the commands to run to actually build a working version of chassis with brief explanation.

MichaelArestad:
Removed git references and simplified the Readme a bit.

README.md
```diff
@@ -11,3 +11,25 @@ Chassis is an attempt at creating open standards designed for CSS libraries, Jav
...
+
+## Getting started
+
+### Prerequisites
+
+Node.js and npm should be installed. [[installer]](http://nodejs.org/download/)
+
+### Building
+
+Once you've cloned Chassis to your machine, run `npm install` and `grunt build` from the root Chassis directory.
```
markelog: `grunt build` - suggest you to have an npm alias for this, like in core - https://github.com/jquery/jquery/blob/master/package.json#L54 then you would never have grunt versions conflicts.

README.md
```diff
@@ -11,3 +11,25 @@ Chassis is an attempt at creating open standards designed for CSS libraries, Jav
...
+
+```
+$ npm install
+$ grunt build
+```
+
+### Other grunt commands
```
markelog: `grunt -h` isn't enough?

Gruntfile.js
markelog: Maybe you could use that alias in here too?

scss/_utilities/_clearfix.scss
```diff
@@ -0,0 +1,19 @@
...
+	&:after {
+		content: "";
+		display: table;
+	}
+	&:after {
+		clear: both;
+	}
+}
+
+.clearfix {
```
gnarf: Seems weird that we'd define one of these.  Do we use it a whole bunch? Doesn't seem necessary in modern design patterns IMO.

Is `overflow: auto` not descriptive enough for the cases where we want `.clearfix`.

Also, is this actually going to define `.clearfix` as a class in the output file? Seems like we'd at least want to "namespace" that?

MichaelArestad: > Do we use it a whole bunch?

I know I do and many others do. We're also supporting IE8 plus currently

`overflow` is definitely what I use when I can, but it can cause unexpected side effects (scroll bars, positioning issues, cut off child elements). They aren't the worst side effects in the world, but they're there. [#](http://bocoup.com/weblog/new-overflow-default/)

It currently does output `.clearfix` in the output file. This can change.

arschmitz: Regarding naming here this really is dependent upon final decision on https://github.com/jquery/css-chassis/issues/4

battaglr:
> That's partially why I might prefer to just not include the compiled file. As we're eventually going to have a build system, there's not a huge need to have the compiled file at all.

I have learned the hard way that including the compiled file will cause some headaches sooner than later.

scss/_utilities/_clearfix.scss
sfrisk: We can leave it in for name (the name), and if we decided to have different naming conventions for #4, then we can change it.

Also might as well leave the clearfix mixin there for now, since in some cases it could definitely be handy to have.

arschmitz:
Closing replaced by #42

Issue #19, opened 2014-11-19
Run performance tests on existing frameworks


scottgonzalez:
Once #2 is complete and we have a performance testing system, we should run existing frameworks through the system to get an idea of what works well and what doesn't. This will help us identify patterns to use and avoid during development.

PR #34, opened 2015-01-20
Performance: Add performance test suite for comparing other frameworks

arschmitz:
This is a start to addressing #19

This adds a css rendering performance framework for testing various css-frameworks using perfjankie
Right now this only adds Bootstrap, Foundation, jQuery UI, and jQuery Mobile and only the button component.

Additional frameworks / components are easily added by creating a a `framework.js` file which returns an object containing objects with hashes of options and a markup generator function see 4 existing frameworks for reference. 

To host the test files a connect server is started that uses EJS templates to generate X number of components.

The results are stored in a couchdb that needs to be separately setup.
current results http://104.236.81.132:5984//css-perf/_design/site/index.html#/page-select


jzaefferer:
> The results are stored in a couchdb that needs to be separately setup.

Is there any way around that?

Otherwise, does it really belong in this repo? Seems like a separate repo with just the performance tests, across various projects, might make more sense. I don't see this same setup landing in both UI and Mobile, for example.

arschmitz:
@jzaefferer there is no way around needing couchDB to be able to view results of tests locally. HOWEVER this is of limited use locally because its so environment dependent. Also setup is a single download and double click no configuration needed.

The setup would likely be similar but not the same for mobile this would run via a CI for consistency and likely use auto generated component pages ( but only from ui or mobile not other frameworks )

rxaviers:
I don't want to slow any plans down. But, I'm :+1: for having comparisons with other frameworks in a separate project. Potentially, such comparison-project could count with collaboration from maintainers and users of the tested frameworks. Analogous (but, hopefully better) to what https://github.com/rxaviers/javascript-globalization/ is for i18n.

performance/component.html
```diff
@@ -0,0 +1,20 @@
...
+		});
+	%>
+	<%- stylesheets %>
+</head>
+<body>
+	<h1><%= title %></h1>
+	<h2>Component Count: <%= count %></h2>
+	<%- html %>
+</body>
+</html>
```
arschmitz: Missing new line at EOF

Gruntfile.js
```diff
@@ -1,21 +1,129 @@
 module.exports = function( grunt ) {
 
 grunt.loadNpmTasks( "grunt-contrib-jshint" );
+grunt.loadNpmTasks( "grunt-contrib-connect" );
+grunt.loadNpmTasks( "grunt-selenium-server" );
 grunt.loadNpmTasks( "grunt-html" );
 grunt.loadNpmTasks( "grunt-jscs" );
 
+var target,
```
gnarf: before this `Gruntfile` gets out of hand, do you think we could adopt the `grunt.loadTasks('build/tasks')` approach http://www.thomasboyt.com/2013/09/01/maintainable-grunt.html

gnarf: also require('load-grunt-tasks')(); ? :)

arschmitz: @gnarf yeah i'm sure i would have switched it eventually. Just as a note however most of this will be going away. I'm working to integrate much of this into perfjankie / browser-perf directly since they are really dependencies of those. Specifically the selenium stuff.

PR #44, opened 2015-03-05
Performance:  Add Setup step ,"Create external directory", to README

geekman-rohit:
After building and before running first performance tests with `grunt perf` , user must create directories where downloaded external files (selenium jar file) will be stored.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ $ grunt build
 ## Setup
 
 * Follow the steps for building above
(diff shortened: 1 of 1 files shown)


arschmitz:
@geekman-rohit Thanks for doing this. The more i thought about it though, I think id prefer to have this automatic, and also keep the server in a different location. If we land #47 that will replace and close this.

PR #46, opened 2015-03-08
Component: Performance FrameWork Fixes, Added Semantic-UI

### LINKED ISSUE #2: Performance testing
We should have performance tests that run on every commit to ensure our styles are not causing rendering issues.

@kristoferjoseph and @brianleroux have experience doing this with Topcoat.

prasunanand:
Added semantic-ui.js and added the task in perfjankie.js
 
Fixes #2 
Ref #2


arschmitz:
@prasunanand this is looking better but it looks like you removed a bunch of the variations you originally had why is this?

prasunanand:
@arschmitz  I removed the social-network buttons cause  they were just not useful and I grouped a bunch of buttons in the misc category.

prasunanand:
@arschmitz  Can you suggest the changes to be made? I would like to add more frameworks to the performance testing and also add components.

PR #49, opened 2015-03-10
Build: Add CI testing

arschmitz:
Adds a test task which runs on travis also add csslint and cssmin,
this also moves all generated css to a dist/css folder and adds a non minified
version of the css as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ node_modules/
 external/
 icons/svg-min/
 .sass-cache/
+css
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,9 @@
+sudo: false
+language: node_js
+node_js:
+  - "0.10"
+cache:
+  directories:
+  - node_modules
+script:
+  - "grunt"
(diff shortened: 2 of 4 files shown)


jzaefferer:
Seems odd that you need `cssmin` when there's already `outputStyle: "compressed"`. Otherwise diff looks good to me.

.gitignore
```diff
@@ -9,3 +9,4 @@ node_modules/
 external/
 icons/svg-min/
 .sass-cache/
+css
```
jzaefferer: Shouldn't this be `dist/`? Or why does #50 change this line?

arschmitz:
I meant to add a note about why i was adding css min here. Its actually two fold.
1.) the compressed style is not as good as css min because it doe not re order and do a couple other minor things the improve copmressability with gzip.
2.) CSS min gives you final gzip size of the files it creates which is super important to know when working with bem, because gzip compressibility is important to keep file size down, since bem tends to bloat badly in terms of LOC but is a huge winner in final gzip'ed file size when done / ordered properly. 

For reference on this see top coat here https://docs.google.com/presentation/d/1_15yz1uT-oq17ZNsZTczGgwUh-TP15gGO6DFcz_XTJU/edit#slide=id.g688356856_20

PR #51, opened 2015-03-10
Performance: Add Dojo Widget to test suite

geekman-rohit:
This PR adds Dojo Widget (dijit) test to the test suite .
[Link to the documentation dijit/form/Button](http://dojotoolkit.org/reference-guide/1.10/dijit/form/Button.html#dijit-form-button)

There are not many variants I could figure out from the documentation. Though there are combo buttons, I have not added those because similar combo-button with dropdown are available in other frameworks like bootstrap, and they were not added in test suite, hence, it would not have been a fair comaprison.

### DIFF
--- a/performance/frameworks/dojo.js
+++ b/performance/frameworks/dojo.js
@@ -24,7 +24,7 @@ module.exports = {
 				button += " dijitDisplayNone";
 			}
 
-			button += " '>cut</span></span></span>" +
+			button += " '>button</span></span></span>" +
 				"<input type='button' value='' class='dijitOffScreen' " +
 				"aria-hidden='true'></span></span>";
 			return button;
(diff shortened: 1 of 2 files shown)


performance/frameworks/dojo.js
```diff
@@ -0,0 +1,49 @@
...
+			button += "'></span>" +
+				"<span class='dijitReset dijitToggleButtonIconChar'>●</span>" +
+				"<span class='dijitReset dijitInline dijitButtonText ";
+
+			// Check if Button text is to be displayed
+			if ( options.type ) {
+				button += " dijitDisplayNone";
+			}
+
+			button += " '>cut</span></span></span>" +
```
arschmitz: you have `cut` hard coded here lets use the icon value or a generic text of `button`

geekman-rohit:
Yes, thanks for pointing out! updated  the PR with a new commit.

arschmitz:
@geekman-rohit your using input type=buttons here instead of button elements can you switch that for consistency with other tests please.

geekman-rohit:
@arschmitz dojo replaces button with the markup generated above, this markup is added by dojo javascript in place of a button element.

arschmitz:
@geekman-rohit ah your right thank you

PR #56, opened 2015-03-15
Kendo ui mobile

### LINKED ISSUE #2: Performance testing

prasunanand:
Added kendo-ui-mobile.js and added the task in perfjankie.js

Fixes #2 
Ref #2

kendo-ui-mobile.js varies with kendo-ui.js:
1. extra cdn
2. all the buttons have a class of 'km-button' instead of 'k-button'


arschmitz:
@prasunanand sorry i think you misunderstood what I meant by add it as another variation, I meant add this as a third option type in the `variations` array
```js
type: [
    "k-button",
    "km-button"
]
```

I'm going to close this since it should just be added to your original PR #55

Issue #16, opened 2014-11-13
Typography


sfrisk:
Assuming we want to cover typography along with form elements in the first release, the following items need to be considered:

What font do we want to use?  Do we want to allow for different fonts, and allow user to choose? Font color?

We need styling (font, font size, line-height, font-weight, etc) for most if not all of the following elements:

* Headers (`<h1>`, `<h2>`, `<h3>`, `<h4>`, `<h5>`, `<h6>`)
* Classes meant to mimic headers, but not semantically mean the header they're mimicking ( For example, a `.h1` class to use on an `<h2>` that someone wants to appear like an `<h1>`.
* Paragraphs (`<p>`)
* Inline text elements (`<mark>`, `<del>`, `<s>`, `<ins>`, `<u>`, `<small>`, `<strong>`, `<em>`, `<b>`, `<i>`)
* Lists (Ordered, Unordered, Definition)
* Block quotes? Addresses?

Also do we want any utility typography classes, for text alignment or text transformation?

GreatDanton:
:+1:  for Source Sans Pro. When I was writing my site, I was searching for font that would look nice no matter at which size. So I was browsing around, looking for sites and marking the fonts that I like. I ended up with 5 fonts. Lato, Open Sans, Droid sans, Roboto, Source Sans Pro. After trying each of them I decided Source Sans Pro got the best word spacing.

Oh, and hi btw :)

PR #57, opened 2015-03-15
Initial base typography

MichaelArestad:
Still in progress. Will keep pecking away at it.

https://github.com/jquery/css-chassis/issues/16

### DIFF
--- a/demos/typography.html
+++ b/demos/typography.html
@@ -2,7 +2,6 @@
 <html>
 <head>
 	<meta charset="utf-8">
-	<meta content="ie=edge">
 	<title>CSS Chassis - Typography</title>
 	<meta name="description" content="Typography skeleton for styling">
 	<meta name="viewport" content="width=device-width, initial-scale=1">
--- a/demos/typography.html
+++ b/demos/typography.html
@@ -1,12 +1,12 @@
 <!doctype html>
 <html>
 <head>
-	<meta charset="utf-8">
-	<meta http-equiv="x-ua-compatible" content="ie=edge">
+	<meta charset='utf-8'>
+	<meta http-equiv='x-ua-compatible' content='ie=edge'>
 	<title>CSS Chasis - Typography</title>
-	<meta name="description" content="Typography skeleton for styling">
-	<meta name="viewport" content="width=device-width, initial-scale=1">
-	<link rel="stylesheet" href="../dist/css/chassis.css">
+	<meta name='description' content='Typography skeleton for styling'>
+	<meta name='viewport' content='width=device-width, initial-scale=1'>
+	<link rel='stylesheet' href='../dist/css/chassis.css'>
 	<link href='http://fonts.googleapis.com/css?family=Source+Sans+Pro:400,600,700,400italic,700italic' rel='stylesheet' type='text/css'>
 	<style>
 		body {
@@ -53,7 +53,7 @@
 
 <h2>Block-level elements</h2>
 
-<p>Here's a paragraph &lt;p&gt; filled with some <a href="http://spaceipsum.com/" target="_blank">Space Ipsum</a>. Never in all their history have men been able truly to conceive of the world as one: a single sphere, a globe, having the qualities of a globe, a round earth in which all the directions eventually meet, in which there is no center because every point, or none, is center — an equal earth which all men occupy as equals. The airman's earth, if free men make it, will be truly round: a globe in practice, not in theory.</p>
+<p>Here's a paragraph &lt;p&gt; filled with some <a href='http://spaceipsum.com/'>Space Ipsum</a>. Never in all their history have men been able truly to conceive of the world as one: a single sphere, a globe, having the qualities of a globe, a round earth in which all the directions eventually meet, in which there is no center because every point, or none, is center — an equal earth which all men occupy as equals. The airman's earth, if free men make it, will be truly round: a globe in practice, not in theory.</p>
 
 <p>There can be no thought of finishing for ‘aiming for the stars.’ Both figuratively and literally, it is a task to occupy the generations. And no matter how much progress one makes, there is always the thrill of just beginning.</p>
 
@@ -66,7 +66,7 @@
 	<p>End of the article.</p>
 </article>
 
-<blockquote>Let's light this fire one more time, Mike, and witness this great nation at its best. &lt;blockquote&gt; via <a href="http://spaceipsum.com/" target="_blank">Space Ipsum</a></blockquote>
+<blockquote>Let's light this fire one more time, Mike, and witness this great nation at its best. &lt;blockquote&gt; via <a href='http://spaceipsum.com/'>Space Ipsum</a></blockquote>
 
 <p>We choose to go to the moon in this decade and do the other things, not because they are easy, but because they are hard, because that goal will serve to organize and measure the best of our energies and skills, because that challenge is one that we are willing to accept, one we are unwilling to postpone, and one which we intend to win.</p>
 
@@ -127,7 +127,7 @@
 
 <ul>
 	<li><a href="#">anchor link</a> &lt;a&gt;</li>
-	<li>abbreviation: <abbr title="Syntactically Awesome Stylesheets">SASS</abbr> &lt;abbr&gt;</li>
+	<li>abbreviation: <abbr title='Syntactically Awesome Stylesheets'>SASS</abbr> &lt;abbr&gt;</li>
 	<li><b>bold</b> &lt;b&gt;</li>
 	<li>citation: <cite>A Study in Scarlet</cite> &lt;cite&gt;</li>
 	<li>Here's the <code>&lt;code&gt;</code> element</li>
@@ -138,8 +138,8 @@
 	<li><ins>inserted text</ins> &lt;ins&gt;</li>
 	<li>keyboard input: <kbd>control + space</kbd> &lt;kbd&gt;</li>
 	<li><mark>mark element</mark> &lt;mark&gt;</li>
-	<li>meter: <meter min="200" max="500" value="423">awesomeness</meter> &lt;meter&gt;</li>
-	<li>progress: <progress value="90" max="100">70 %</progress> &lt;progress&gt;</li>
+	<li>meter: <meter min='200' max='500' value='423'>awesomeness</meter> &lt;meter&gt;</li>
+	<li>progress: <progress value='90' max='100'>70 %</progress> &lt;progress&gt;</li>
 	<li>quotation (short): <q>You just gotta believe!</q> &lt;q&gt;</li>
 	<li>quotation within a quotation (short): <q>Yo dawg, I heard you like quotes so I put <q>this quote</q> within this quote.</q> &lt;q&gt;</li>
 	<li>strikethough: <s>Lance Armstrong remains steroid-free.</s> &lt;s&gt;</li>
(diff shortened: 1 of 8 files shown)


typography.html
```diff
@@ -0,0 +1,158 @@
+<!doctype html>
```
sfrisk: Do we want to include this as part of the commit? A page like this would probably count as being part of the "Pattern Library," and probably shouldn't be kept in the home directory

scss/atoms/typography/_typography.scss
```diff
@@ -5,3 +5,120 @@
 @import
 	"variables",
 	"functions";
+
+// TODO probably should do the import in the `head` in case character sets need to change dynamically
+@import url( http://fonts.googleapis.com/css?family=Source+Sans+Pro:400,600,700,400italic,700italic );
```
arschmitz: This results in an `@import` in the final min file and is against the style guide. This should just be an additional link in the `.html` file.

scss/atoms/typography/_typography.scss
```diff
@@ -5,3 +5,120 @@
...
+h2,
+h3,
+h4,
+h5,
+h6 {
+	font-weight: 600;
+	line-height: 1;
+}
+
+h1 {
```
arschmitz: Heading elements should only be defined once.

scss/atoms/typography/_typography.scss
```diff
@@ -5,3 +5,120 @@
...
+// ==========================================================================
+
+a {
+	color: inherit;
+	text-decoration: none;
+	background-image: linear-gradient( to bottom, rgba( 0, 0, 0, 0 ) 75%, #000 50% );
+	background-repeat: repeat-x;
+	background-size: 2px 2px;
+	background-position: 0 em( 22px );
+	outline: 0;
```
arschmitz: Outlines should only be modified using :focus. Use of outline: none or outline: 0 should be limited to :focus rules. (outline-none)

arschmitz:
Now that there is an html file we need to add the `html` task to add html linting. It's already in `package.json`

sfrisk:
Not to mention we need to rework the HTML style guide.

On Tue, Mar 17, 2015 at 11:55 AM, Alexander Schmitz <
notifications@github.com> wrote:

> Now that there is an html file we need to add the html task to add html
> linting. It's already in package.json
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jquery/css-chassis/pull/57#issuecomment-82432647>.
>



-- 
Sarah Frisk
sarahfrisk.com

Twitter: twitter.com/SarahFrisk
LinkedIn: linkedin.com/in/sarahfrisk
Github: github.com/sfrisk

scss/atoms/typography/_typography.scss
MichaelArestad: Why? I'm not opposed to the change, but I'm curious as to the reason.

arschmitz: Its a pretty big accessibility issue because the outline property is what makes the focus highlight in most browsers. So if you set it to `0` on all `a` elements now there is no focus indication on any links on the page. Here is a good explination in detail if your interested. https://github.com/CSSLint/csslint/wiki/Disallow-outline%3Anone

MichaelArestad: I'm aware, but that's a non-issue if there are other visual focus styles in place, which there are.

arschmitz: The only focus style i see here is setting a background size and position which does not really do anything besides make the underline slightly bolder unless i'm missing something? I consider that a pretty big accessibility issue.

This can be seen tabbing through your demo too.  I don't think think people would even notice the focus indicator here in general.
http://view.css-chassis.com/typography/typography.html
Compare this to the focus style on a normal link without a styles. 
http://jsbin.com/waxeyo/1

I personally am against changing the focus outline at all from what the native is.

arschmitz: Just to give some context on why i don't think we should alter native. We used to have non native focus styles in jQuery UI and had TONS of complaints, until we removed it.
Based of the accessibility concerns involved with focus indication i think its best to leave the style users are familiar with on their particular device. 

I do however also think we should provide a visual focus style for certain situations like showing a focus indication where native can't an example is toolbars or accessible styled radio buttons and checkboxes. This situation came up recently and after consulting with several accessibility experts this is what we came up with. 
http://view.jqueryui.com/button-classes/demos/checkboxradio/default.html
You can see tabbing a visual focus outline ( that may depending on your browser / os actually look native )

The problem is the visual button is a label. Labels can't receive focus they transfer to the underlying input which is accessibly hidden. So this means there is no focus outline visible. So we created a class that adds a visual focus indicator to the label. Problem here is the visual focus style and the regular focus style can be inconsistent on some platforms. 

We add our own focus style everywhere for this same reason in jQuery mobile and this solves the issue of visual inconsistency however not the issue of user complaints ( or my own person desire to keep things as simple as possible for end users, navigating through a page via keyboard, by keeping important visual indicators native, as the way people are familiar with, in style ).

arschmitz:
oops wrong number reopening

arschmitz:
I pushed the latest update on this to `view.css-chassis.com`

demos/typography.html
```diff
@@ -0,0 +1,159 @@
+<!doctype html>
+<html>
+<head>
+	<meta charset="utf-8">
+	<meta http-equiv="x-ua-compatible" content="ie=edge">
+	<title>CSS Chasis - Typography</title>
+	<meta name="description" content="Typography skeleton for styling">
+	<meta name="viewport" content="width=device-width, initial-scale=1">
+	<link rel="stylesheet" href="../dist/css/chassis.css">
+	<link href='http://fonts.googleapis.com/css?family=Source+Sans+Pro:400,600,700,400italic,700italic' rel='stylesheet' type='text/css'>
```
arthurvr: single quotes

MichaelArestad:
Updated a few things. It's a good base that's ready to merge IMO. We can then iterate, add colors, etc, in other PRs.

demos/typography.html
arthurvr: no what I meant is that this tag uses single quotes and all the rest doubles. [We use double quotes](http://contribute.jquery.org/style-guide/html/).

MichaelArestad: Ha! Will change.

MichaelArestad:
Failed check because of the hyphen in the attribute name, BTW.

package.json
```diff
@@ -59,6 +59,7 @@
     "grunt-csscomb": "3.0.0",
     "grunt-git-authors": "2.0.0",
     "grunt-html": "1.6.0",
+    "grunt-htmllint": "^0.2.7",
```
arschmitz: We use firm versions to do our best to avoid weird issues with different people having different versions.

arschmitz:
I'm wondering if we really want the `x-ua-compatible` meta since this forces the browser mode. while on production sites i would always use this because the demos are as much for our own visual testing as for demos i think i would prefer to leave it out. 

Also just as a side note, we have never included this on the ui or mobile demos and there has never been an issue. We do of course keep it on our production WP sites though.

arschmitz:
Can you also add demos.css to the csslint settings?

package.json
MichaelArestad: Cool. Will change.

MichaelArestad:
@arschmitz Changes made:

* added hard version.
* removed `x-ua-compatible`
* added `demos.css` to `csslint`

demos/typography.html
```diff
@@ -0,0 +1,151 @@
+<!doctype html>
+<html>
+<head>
+	<meta charset="utf-8">
+	<meta content="ie=edge">
```
arthurvr: If we remove the `x-ua-compatible` there's no deal with including this anymore.

demos/typography.html
```diff
@@ -0,0 +1,151 @@
...
+	<title>CSS Chasis - Typography</title>
+	<meta name="description" content="Typography skeleton for styling">
+	<meta name="viewport" content="width=device-width, initial-scale=1">
+	<link rel="stylesheet" href="../dist/css/chassis.css">
+	<link rel="stylesheet" href="demos.css">
+	<link href="http://fonts.googleapis.com/css?family=Source+Sans+Pro:400,600,700,400italic,700italic" rel="stylesheet" type="text/css">
```
arthurvr: Remove the `type="text/css"`

demos/typography.html
```diff
@@ -0,0 +1,151 @@
+<!doctype html>
+<html>
+<head>
+	<meta charset="utf-8">
+	<meta content="ie=edge">
+	<title>CSS Chasis - Typography</title>
```
arthurvr: Typo `Chassis`

sfrisk:
Any more updates on this?

MichaelArestad:
@sfrisk Good to go.

scss/atoms/typography/_mixins.scss
```diff
@@ -0,0 +1,26 @@
+// ==========================================================================
```
arschmitz: Sorry i didn't notice the before but the style guide mandates `/* comment */` style comments since these are the only valid style in normal css. @MichaelArestad can you please switch all the comments here to use this style.

MichaelArestad: Is using multiline comments the plan for the Sass guide?

arschmitz: I just realize this style has slipped in already in other files, we should switch these to match the style guide in a separate PR

arschmitz: @MichaelArestad well its not necessarily multi line it the only valid style in normal css. I personally don't think its good practice to use different comment styles in different files so we should stick to valid css comments i don't think using `//` has any benefit over `/* */` other then not typing 2 chars which git stripped out in real build anyway i think clarity and consistency should prevail here. @sfrisk what do you think?

MichaelArestad: There are minor benefits which are situational. No single line comments end up in output CSS files whereas multiline comments do depending on output style. This is handy for when you want to comment things like functions without it ending up in the output code or when you are generating a theme file and want to curate which comments end up in the compiled CSS file.

sfrisk: I'll admit, I use `//` in my personal projects using SCSS.  I think we should be consistent with our use of comments structure.  Either use all `//` or `/* */`, or else define specific use cases for each version.

Advantage of `//` is it is similar to our JavaScript style guide, and could help promote consistent commenting styling across all JS and SCSS files. http://contribute.jquery.org/style-guide/js/#comments

That being said, if we're trying to make sure our SCSS stays as close to valid CSS has possible, then we should stick to `/* */`

arschmitz: > No single line comments end up in output CSS files whereas multiline comments do depending on output style
So no comments at all end up in minified files which is really what matter. We also use the "Expanded" style which outputs both single and multi line comments so there is no difference here either. 

> This is handy for when you want to comment things like functions without it ending up in the output code or when you are generating a theme file and want to curate which comments end up in the compiled CSS file. 

This would be true if our output style accommodated this however it does not. Only the compressed output style accommodates this

MichaelArestad: Expanded does not output single line comments. No style does.

Compressed, expanded, and nested support outputting multiline comments (without doing a forced comment `/*! */`).

arschmitz: @MichaelArestad Your right sorry i looked this up and misread. You are correct here.

So in this case there is basically one major case that makes sense to use single line comments which is where we have a comment that would end up with no context and make no sense if it ended up in an uncompressed output file. This would likely be the case for most functions. Most other places should probably stick to multiline i think still. Since we have both css and scss files in this repo i would like to try to keep things as clear and consistent as possible.

@MichaelArestad how does that sound to you?

MichaelArestad: Yeah. Probably functions and comments that are implementation instructions for mixins as well. Will change on another PR.

arschmitz: > implementation instructions for mixins as well

These should probably eventually move to the documentation ( once such a thing exists ) but for now yes i agree.

PR #65, opened 2015-03-23
Performance: add Gumby framework and Skeleton to test suite

ramasilveyra:
This PR adds Gumby framework and Skeleton to the performance test suite.

### DIFF
--- a/performance/frameworks/skeleton.js
+++ b/performance/frameworks/skeleton.js
@@ -1,0 +1,17 @@
+module.exports = {
+	css: [
+		"//cdnjs.cloudflare.com/ajax/libs/skeleton/2.0.4/skeleton.min.css"
+	],
+	button: {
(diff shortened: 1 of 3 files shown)


performance/frameworks/skeleton.js
```diff
@@ -0,0 +1,17 @@
+module.exports = {
+	css: [
+		"//cdnjs.cloudflare.com/ajax/libs/skeleton/2.0.4/skeleton.min.css"
+	],
+	button: {
+		generator: function( options ) {
+			var button = "<button class='" + options.state + "'>";
```
arschmitz: looks like your missing the class `button` here

arschmitz:
I have some doubts about what value skeleton framework adds since its SO minimal only one button variation essentially just that and some typography and a grid. Gumby seems good though.

PR #70, opened 2015-04-10
Colors: Adding Chassis palette and default colors

sfrisk:
Added a color palette file where we can define palette variables.  The theme roller could potentially edit this file with a custom palette and then adjust the variable values in _colors.scss accordingly. 

Edited the _colors.scss file to include some base color variables that could goes along with the typography PR, with appropriate palette variables assigned it it.  Commented out success/error/warning variables are there mostly because I think this ties in with validation styling, and I wasn't quite sure what colors in the current palette would work with these variables, or if we should add additional colors for these cases.  I am open to suggestions on this point.  That or I could just remove those variable declarations for now and we could address them when we get to #9.


sfrisk:
Upon further consideration, tweaking the palette names to be a bit more generic, to allow for an easy demo file of the palette.

scss/_utilities/_colors.scss
```diff
@@ -1,3 +1,14 @@
...
 // ==========================================================================
+
+
+$color-background: $gray-light;
+$color-font: $gray-dark;
+$color-link: $blue-dark;
+$color-button: $blue;
+$color-button-text: $gray-light;
+
+// $color-success: $blue;
```
MichaelArestad: You might suggest naming these something like `$color-alert-yellow`.

sfrisk:
redoing this in a different way.

PR #87, opened 2015-06-04
Typography: Remove overwriting of anchor elements

### LINKED ISSUE #81: Typo on slim-underline() mixin
Hello,

There's a typo on your `slim-underline()` mixin.

The mixin says `@mixin slim-underline()` but the 'usage' example in the comment says `// Usage: @include slime-underline;`

_slim_ on one and _slime_ on the other.

https://github.com/jquery/css-chassis/blob/master/scss/atoms/typography/_mixins.scss

### LINKED ISSUE #84: Remove the overwriting of the default <a> functionality
I thought about adding to the existing discussion on closed issue #38 but I didn't want it to be missed. 

Quoting a reply in the discussion.

>Phase one of this project is definite more on the CSS framework side of things. Just the stuff you need to get started on a project, with some useful helper classes.

This doesn't seem accurate.

Looking at the code here it seems that the project is already drifting away from basic rules to opinionated design that will have a marked effect on both the ability of developers to overwrite basic styles and the rendering performance within browsers.

https://github.com/jquery/css-chassis/blob/64a51ac31474f7b2c413f8d1767ca776f76228fd/scss/atoms/typography/_typography.scss

Take the code and mixin that is referenced [here](https://github.com/jquery/css-chassis/blob/64a51ac31474f7b2c413f8d1767ca776f76228fd/scss/atoms/typography/_typography.scss#L110)

    p > a {
	// Good example of slim underline. We'll want to use this on body copy/quotes
	color: inherit;
	@include slim-underline;
    }

There's now a global selector with a specificity of 0,0,2  that will affect every page the code is added to. It's overwriting the default styles applied to an `a` tag with behaviour reduces visibility of the link, requires more processing power, and includes hacks for webkit browsers.

https://github.com/jquery/css-chassis/blob/64a51ac31474f7b2c413f8d1767ca776f76228fd/scss/atoms/typography/_mixins.scss

Is this really the intended direction of the project? It seems not.

> Chassis is an attempt at creating open standards designed for CSS libraries, JavaScript UI libraries, and web developers in general. This project will define standards for markup and class names for common widgets and patterns, as well as an implementation of those standards. Just like DOM libraries can build upon Sizzle, we hope UI libraries will build upon this.

sfrisk:
Fixes #84
Fixes #81

### DIFF
--- a/scss/atoms/typography/_mixins.scss
+++ b/scss/atoms/typography/_mixins.scss
@@ -3,24 +3,3 @@
 // ==========================================================================
 
 // Adds a single pixel underline (so nice) on retina
-// Usage: `@include slime-underline;`
-@mixin slim-underline() {
-	text-decoration: none;
-	background-image: linear-gradient( to bottom, rgba( 0, 0, 0, 0 ) 75%, #000 50% );
-	background-repeat: repeat-x;
-	background-size: 2px 2px;
-	background-position: 0 em( 22px );
-
-	&:hover,
-	&:active,
-	&:focus {
-		background-size: 2px 6px;
-		background-position: 0 em( 18px );
-	}
-	@media not all,
-	only screen and ( min-resolution: 2dppx ),
-	only screen and ( -webkit-min-device-pixel-ratio: 2 ) {
-		background-image: linear-gradient( to bottom, rgba( 0, 0, 0, 0 ) 75%, #000 75% );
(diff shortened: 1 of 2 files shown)


sfrisk:
merged it, closing

sfrisk:
merged this, closing

Issue #27, opened 2014-12-17
Create Logo


arschmitz:
Now that we have a name we really need a logo.

@isaacdurazo you had expressed interest in helping do you think this is something you could tackle? It would be a huge help to be able to get things going!

isaacdurazo:
Just FYI guys, I've started to brainstorm a little bit on this. Nothing to show yet but wanted to let you know that I've started working on this.

<3

arschmitz:
@isaacdurazo Awesome thanks for the update! Can't wait!

sfrisk:
Yay!! :+1:

isaacdurazo:
Hey guys, I've been working on this for severals days now and here's a first option. 

I'm keeping it simple by taking the idea of the "chassis" and do something that evokes it a little bit.

![screen shot 2015-01-26 at 10 32 48 pm](https://cloud.githubusercontent.com/assets/1379244/5914042/568e0778-a5ab-11e4-80fc-4b1c5f454e14.png)

What do you think?

sfrisk:
Ooooo.

I like the b/w version and the font, and the overall simplicity of the design. Not quite sure what I think of the logo with the green background, but I like the green with white background.

Overall I like the direction you're going with this though.

arschmitz:
Overall I agree with sfrisk I like the font a lot and the logo. I like the first 2 more then the third also not sure on the green back. 

One comment I would make as far as evoking the idea of a chassis. I get the idea completely of the c being like the internal structure or "chassis". I guess to me though just due to common usage being generally one of 2 specific things ( a car frame or computer case ) it does not really evoke the idea of a chassis to me. That being said I would not necessarily change, to make it evoke that. I like the logo a lot and think it will work well in usage and like the simplicity. Just wanted to throw that out there. 

Nice work though really like it and the font a lot!

isaacdurazo:
Thanks guys, I'm glad you liked it. I was thinking on the idea of the "chassis" more as a base/structure of something, so I didn't want to go to literal there.

Regarding colors, all these 3 are part of the same option. I forgot to mention that part (oops). We need to have the ability to use the logo in different scenarios, like on a white background or a color background, and also a monochrome version. 

You guys mentioned you liked the colors, would you be interested in seeing a few more options? I wanted to show what I had as the graphic before experimenting with colors, that's why I didn't posted other variations.

sfrisk:
[...]

PR #97, opened 2015-07-17
Buttons: Initial pass at buttons, covering sizes, options, disabled, active, focus, and hover states

sfrisk:
This is my first initial pass at buttons.  It this is the direction people want to go with for buttons, I can add a few more options for buttons  (colors, button group, block button etc).


cvrebert:
Why so much `@extend`ing? [It's arguably an antipattern.](https://tech.bellycard.com/blog/sass-mixins-vs-extends-the-data/)

sfrisk:
@cvrebert As I think about it, mixins would probably have be better solution, especially if we want to add additional options.  I can easily switch to using mixins instead. Part of this issue was also to test the BEM naming convention (since we haven't had any commits with that yet), and get an idea if visually this was the type of button we wanted to go with.  I'll update the PR (although probably not tonight) with a mixin version.

sfrisk:
Or you know, maybe just make things mixins tonight, since I can't sleep.  Let me know what you think, and if you approach works.

demos/buttons.html
```diff
@@ -0,0 +1,104 @@
...
+<h2>Button tag</h2>
+
+<p>Use the button classes on an <code>&lt;a&gt;</code>, <code>&lt;button&gt;</code>, or <code>&lt;input&gt;</code> element.
+
+
+<form>
+	<a class="ui-button--default" href="#" role="button">Link</a>
+	<button class="ui-button--default" type="submit">Button</button>
+	<input class="ui-button--default" type="button" value="Input">
+	<input class="ui-button--default" type="submit" value="Submit">
```
cvrebert: Missing a `type="reset"` example

demos/buttons.html
```diff
@@ -0,0 +1,104 @@
...
+<h1>CSS Chassis</h1>
+
+<h2>Button tag</h2>
+
+<p>Use the button classes on an <code>&lt;a&gt;</code>, <code>&lt;button&gt;</code>, or <code>&lt;input&gt;</code> element.
+
+
+<form>
+	<a class="ui-button--default" href="#" role="button">Link</a>
+	<button class="ui-button--default" type="submit">Button</button>
```
cvrebert: Include a `<button type="button">` for completeness?

demos/buttons.html
```diff
@@ -0,0 +1,105 @@
+<!doctype html>
+<html>
+<head>
+	<meta charset="utf-8">
```
cvrebert: Add `<meta http-equiv="X-UA-Compatible" content="IE=edge">` ?

cvrebert:
Have you considered:
* Block-level buttons
* A class to make a `<button>` look like a normal `<a>`
* Additional classes to emulate the appearance of `:active`, `:focus`, :hover`

scss/atoms/buttons/_mixins.scss
```diff
@@ -0,0 +1,64 @@
...
+@mixin ui-button($color, $bgcolor, $disabled: false) {
+	background: $bgcolor;
+	border: 0;
+	color: $color;
+	display: inline-block;
+	font-weight: 500;
+	margin: .25em;
+	text-align: center;
+	text-decoration: none;
+	text-transform: uppercase;
```
cvrebert: Too opinionated, imho

sfrisk:
Block-level buttons were actually next on my list of things to add.  
Making a button look like a link is a good one, I will add it to the list.  Same with classes for active/hover/focus.

scss/atoms/buttons/_mixins.scss
```diff
@@ -0,0 +1,64 @@
...
+	&:active {
+		background: darken($bgcolor, 20%);
+	}
+	&:focus {
+		background: darken($bgcolor, 15%);
+	}
+	&:hover {
+		background: darken($bgcolor, 10%);
+	}
+	&:disabled {
```
cvrebert: Chassis' browser compatibility doesn't seem to be documented, but FYI `:disabled` is not IE8 compatible.

scss/atoms/buttons/_mixins.scss
```diff
@@ -0,0 +1,64 @@
...
+	color: $color;
+	display: inline-block;
+	font-weight: 500;
+	margin: .25em;
+	text-align: center;
+	text-decoration: none;
+	text-transform: uppercase;
+
+	@if($disabled) {
+		cursor: default;
```
cvrebert: Not `cursor: disabled` ?

scss/atoms/buttons/_mixins.scss
sfrisk: Once the jsass PR lands I will probably make that line into one of the configurable variables that the theme roller can set, for those who prefer not to have uppercase buttons.

scss/atoms/buttons/_mixins.scss
sfrisk: That was being covered in normalize.css

scss/atoms/buttons/_mixins.scss
```diff
@@ -0,0 +1,64 @@
+/*
+* ==========================================================================
+* Buttons (Mixins)
+* ==========================================================================
+*/
+
+@mixin ui-button($color, $bgcolor, $disabled: false) {
```
cvrebert: Consider:
```
vertical-align: middle;
touch-action: manipulation;
-VENDORS-user-select: none;
```

scss/atoms/buttons/_mixins.scss
cvrebert: Huh?

scss/atoms/buttons/_mixins.scss
```diff
@@ -0,0 +1,64 @@
...
+	text-transform: uppercase;
+
+	@if($disabled) {
+		cursor: default;
+		opacity: .6;
+	}
+}
+
+@mixin ui-button-sizes($size: default) {
+	@if($size == large) {
```
cvrebert: Perhaps use a Sass map lookup instead? E.g.
```sass
font-size: map-get($button-heights, $size);
line-height: map-get($button-heights, $size);
```

scss/atoms/buttons/_mixins.scss
cvrebert: Also, consider adding `&[aria-disabled="true"]`, which makes toggling the disabledness of an `<a>` button much simpler (since it's context-insensitive and uniform), and also communicates the disabledness to accessibility tools.

On a tangentially related note, perhaps you jQuery folks would consider voicing support for https://www.w3.org/Bugs/Public/show_bug.cgi?id=28673 ?

scss/atoms/buttons/_mixins.scss
```diff
@@ -0,0 +1,64 @@
...
+	}
+	&:focus {
+		background: darken($bgcolor, 15%);
+	}
+	&:hover {
+		background: darken($bgcolor, 10%);
+	}
+	&:disabled {
+		background: $bgcolor;
+		opacity: .6;
```
cvrebert: Note: Not IE8 compatible

cvrebert: Value is repeated on line 20. Perhaps use a variable to DRY things up?

cvrebert:
Consider implementing https://github.com/twbs/bootstrap/pull/15947 as well

scss/atoms/buttons/_mixins.scss
sfrisk: Oh wait, nevermind via email I thought you were commenting somewhere else, I shall adjust.

scss/atoms/buttons/_mixins.scss
sfrisk: Good idea.

scss/atoms/buttons/_buttons.scss
```diff
@@ -0,0 +1,102 @@
...
+* ==========================================================================
+*/
+@import "mixins";
+
+/* These colors won't remain here, will be removed with jsass  */
+$background: #fff;
+$chassis-blue: #4fc0c8;
+$chassis-gray-light: #f2f2f2;
+$chassis-gray-dark: #383838;
+
```
kristyjy: I'm wondering if we should name the variables based on their purpose rather than color? This way when they get changed in the theme roller a person wouldn't create a purple button that is called "blue".

sfrisk: Yeah, I should switch these to stuff like $button-primary-background, and $button-primary-color.  That was something to put in their temporarily, but you're right, I should have better variable names.

cvrebert:
Also, what about combinations of multiple states? (Example: https://github.com/twbs/bootstrap/issues/16767)

sfrisk:
@cvrebert Ooo, I hadn't thought of that one. Might need to tweak how disabled styling currently works to make that work.

sfrisk:
Also @geekman-rohit pointed out that the primary button contrast don't actually pass accessibility contrast checking.  I used colors suggested for use with the [Chassis Logo](https://github.com/jquery/css-chassis/issues/27#issuecomment-90734881), but if anyone has any alternative color suggestions that are accessible, I would love to hear them.

geekman-rohit:
Basically we want the background - foreground colors to have more contrast so they are easier to read to all people. We also need to see what tests it should pass? I think the primary button should pass AAA while it should be okay for disabled buttons to be AA or even A http://www.w3.org/TR/UNDERSTANDING-WCAG20/conformance.html#uc-levels-head

arschmitz:
there are automated tools we can use with grunt for this sort of checking its been on my todo for a while to set them up

sfrisk:
I'll add an issue for Accessibility Checking

sfrisk:
This was brought up in the meeting today, but should we have the default button be a "block" button, since we'll be designing this mobile first, with an optional modifier to have an inline button?  Or the other way around?

geekman-rohit:
I think it should be inline with a modifier to make it block.

sfrisk:
@geekman-rohit added some different types of buttons to test the performance stuff you're working on

sfrisk:
Closing this in favor of #138, which is our new approach on this.

PR #103, opened 2015-07-22
Build: make JSCS and JSHint lint the grunt config too

arthurvr:
This makes JSHint and JSCS lint the files in `tasks/` too. I asked this but seems like @arschmitz forgot it: https://github.com/jquery/css-chassis/pull/89#discussion_r32540343.


arthurvr:
The CI failure is on master too.

tasks/variables.js
```diff
@@ -1,12 +1,12 @@
 module.exports = function( grunt ) {
 	grunt.registerTask( "buildVariables", function() {
-		var jsass = require( "jsass-vars" );
-		var path = require( "path" );
-		var varPath = path.join( process.cwd(), "scss/variables/*" );
+		var jsass = require( "jsass-vars" ),
```
arschmitz: One var is no longer a requirment but we need to update our version of JSCS for it to not complain.

arschmitz:
@arthurvr looks good other then id prefer to update and not use one var any more.

arthurvr:
> not use one var any more.

I considered it unrelated to this PR, but yeah, updated now.

> other then id prefer to update

Can you elaborate?

arschmitz:
@arthurvr I just meant i would prefer to update jscs then leave one var. Thats all i have.

robbbz:
lgtm modulo the ci failure

sfrisk:
Hey @arthurvr, when you have a sec, could you update this with master?

arthurvr:
Yep, I totally can! It'll have to wait ~2 weeks though. It's only a
semicolon.

sfrisk:
Oh if it is just that I will fix it. I didn't look to closely at what was causing it last night.

PR #105, opened 2015-07-25
Linting:Add scss variable files to linting

geekman-rohit:
Adds jsass-vars scss variable files from scss/varibales/ folders to js linting.


scss/variables/chassis.js
```diff
@@ -1,6 +1,6 @@
 // This is just boiuler plate dont worry about it
 // This just supports loading in any enviroment
-(function ( root, factory ) {
+( function( root, factory ) {
     if ( typeof define === "function" && define.amd ) {
         define( [], factory );
```
geekman-rohit: Looks like this file has wrong indentation, skipped my notice, I'll fix that

geekman-rohit:
@sfrisk , updated the PR

tasks/options/jscs.js
```diff
@@ -1,3 +1,6 @@
 module.exports = {
-	all: [ "*.js", "performance/**/*.js" ]
-};
\ No newline at end of file
+	all: [ "*.js", "performance/**/*.js", "scss/variables/**/*.js" ],
+	options: {
+		fix: true
+	}
```
geekman-rohit: I ran it once with fix true no issues. We could get rid of the option now or do we want to keep it ? @arschmitz ?

scss/variables/template.js
```diff
@@ -1,13 +1,14 @@
 // This is just boiuler plate dont worry about it
```
sfrisk: spelling error - "boiler plate" not "boiuler plate"

sfrisk: (This could easily be fixed in a different issue though)

tasks/options/jscs.js
sfrisk: Per discussion with @arschmitz, fix should not be in the repo at all.

scss/variables/template.js
geekman-rohit: its a small thing, ill do it in this PR.

tasks/options/jscs.js
geekman-rohit: yep that shouldn't be here, I actually forgot to remove it. Thanks for reminding.

geekman-rohit:
@sfrisk updated.

PR #114, opened 2015-08-04
Build: add accessibility testing

### LINKED ISSUE #99: Add Automated Accessibility Testing

dylanb:
Fixes #99


arschmitz:
@dylanb feel free to force push PR's instead of opening new ones

dylanb:
wooo hooo! Build passed.

tasks/options/axe-webdriver.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+	firefox: {
```
arschmitz, line 3: Why only firefox? We do chrome and firefox for the performance testing id like to do the same here if possible

Gruntfile.js
```diff
@@ -1,10 +1,11 @@
 module.exports = function( grunt ) {
-var path = require( "path" );
-require( "load-grunt-config" )( grunt, {
-	configPath: [
-		path.join( process.cwd(), "tasks/options" ),
-		path.join( process.cwd(), "tasks" )
-	],
-	init: true
-} );
+	grunt.loadNpmTasks("axe-grunt-webdriver");
```
arschmitz: Is this actually needed? `load-grunt-config` should load it.

tasks/options/axe-webdriver.js
dylanb, line 3: Chrome does not work on Travis. I could add a chrome target for use locally.

Gruntfile.js
```diff
@@ -1,10 +1,11 @@
...
+	var path = require( "path" );
+	require( "load-grunt-config" )( grunt, {
+		configPath: [
+			path.join( process.cwd(), "tasks/options" ),
+			path.join( process.cwd(), "tasks" )
+		],
+		init: true
+	} );
```
arschmitz: Why the indentation change on all of this? Can you please outdent it back to where it was to follow our style guide.

Gruntfile.js
dylanb: It does not load it - same as perfjankie - I could add it to the options?

tasks/options/axe-webdriver.js
arschmitz, line 3: Could you please. We use browserstack for cross browser testing on travis. Can axe-webdriver use browserstack? We don't run the performance stuff on browserstack because browserstack is not consistent enough for performance testing.

tasks/options/axe-webdriver.js
```diff
@@ -0,0 +1,19 @@
...
+grunt.loadNpmTasks("axe-grunt-webdriver");
+return {
+	firefox: {
+		options: {
+			threshold: 0
+		},
+		urls: ['http://localhost:4200/demos/typography.html'],
+		dest: 'tmp/gu.json'
+	},
+	chrome: {
```
geekman-rohit: Didn't we want to remove chrome?

arschmitz: why would we not want chrome?

tasks/options/axe-webdriver.js
rohmulan, line 3: Adding chrome will crash travis ?. Maybe we could have two tests, one for travis and another with chrome to run locally.

arschmitz:
@dylanb my comment got lost when you updated but can axe-webdriver run on browserstack?

tasks/options/axe-webdriver.js
geekman-rohit: The travis will crash otherwise. We should have a separate test that uses chrome, that test will not be a part of travis.

arschmitz: we should not be running this at all on travis directly it should be on browserstack.

dylanb:
@arschmitz there is no reason it should not run on browserstack - I have never set it up to run there - how do the browserstack tests run for chassis?

arschmitz:
@dylanb there are none for chassis yet this is the first thing that would use it. Sorry i used "we" a bit ambiguously here it was more a general "jquery foundation" we. We have a browserstack account for all our CI cross browser testing so we should use that here as well.

dylanb:
@arschmitz could we handle that as a separate issue? I will need some help getting browserstack set up.

arschmitz:
@dylanb thats fine if its not a part of axe-webdriver already its probably going to need to be done there first i can probably help with that.

dylanb:
@arschmitz that would be really awesome!

Gruntfile.js
arschmitz: Hmm figured out the cause of this and opened an issue on `axe-grunt-webdriver` for a possible fix

by default plugins to load grunt tasks look for the pattern `grunt-*`

dylanb: I moved it to the options file and once the plugin is updated, we can remove it

dylanb:
@arschmitz I have renamed the axe grunt Webdriver plugin to grunt-axe-webdriver and made the appropriate changes to this PR. What else do I need to do to get this merged?

sfrisk:
@arschmitz @dylanb 
Just checking to see what the current status of this PR is, and if there is anything that needs to be done on our end to get everything working.

dylanb:
@sfrisk I am waiting on feedback from someone as to what I need to do to make this mergeable...bit-rot has caused it to get out of sync with other code changes but I can fix that quite quickly

sfrisk:
Hey, @dylanb, I think this looks good (sorry about the late response), could you just merge latest master into this to make sure everything still runs?  If it does, this PR should be good to go.

geekman-rohit:
hey @dylanb updates?
@sfrisk I think we should pull this is in before #140

sfrisk:
@geekman-rohit I've been working on pulling it in, but there are some merge conflicts I need to deal with.

PR #111, opened 2015-07-31
Build: heirarchial prefixer for scss variable files

geekman-rohit:
This file will help with prefixing for jsass-vars variable files. It also allows heirarchial prefixing, allowing us to create strucuture in variable files. 
eg.
```javascript
chassis.meter = {
	'ui-meter-height': { 
		name: ' Meter Height',
		value: '24px' 
	},
	'ui-meter-width': {
		 name: 'Meter Width',
		 value: '200px' 
	},
	'ui-meter-border-radius': {
		 name: 'Meter Border Radius',
		 value: 0 
	},
	'ui-meter-border-size': {
		 name: 'Meter Border Size',
		 value: '1px' 
	},
	'ui-meter-border-color': {
		 name: 'Meter Border Color',
		 value: '$color-gray-dark' 
	} 
}
```
Using this module we could do : 
```javascript
var meter = {
ui: {
		meter: {
			height: {
				name: " Meter Height",
				value: "24px"
			},
			width: {
				name:"Meter Width",
				value:"200px"
			},
			border:{
				radius: {
					name: "Meter Border Radius",
					value: 0
				},
				size: {
					name: "Meter Border Size",
					value: "1px"
				},
				color: {
					name: "Meter Border Color",
					value: "$color-gray-dark"
				}
			}
		}
	}
}
chassis.meter =  prefixer( meter );
```
We could include it if it could be helpful.

PR #116, opened 2015-08-10
Progress and Meter styling

rohmulan:
Some basic styling for progress bar. Still working on it. 

I would love some suggestions on making them visually look different a bit, I setup a [codepen](http://codepen.io/geekman-rohit/pen/qOGKMK)
Would like suggestions on a better file organization too.

### DIFF
--- a/demos/progress.html
+++ b/demos/progress.html
@@ -1,0 +1,125 @@
+<!doctype html>
+<html>
+<head>
+	<meta charset="utf-8">
+	<title>CSS Chassis - Typography</title>
+	<meta name="description" content="Typography skeleton for styling">
+	<meta name="viewport" content="width=device-width, initial-scale=1">
+	<link rel="stylesheet" href="../dist/css/chassis.css">
+	<link rel="stylesheet" href="demos.css">
+	<link href="http://fonts.googleapis.com/css?family=Source+Sans+Pro:400,600,700,400italic,700italic" rel="stylesheet">
+</head>
+<body>
+
+<h1>CSS Chassis</h1>
+<h2>Progress and Meter</h2>
+<hr>
+<h3>Meter Element</h3>
+<meter max="100" value="70" min="0" low="50" high="90"></meter>
+<meter max="100" value="40" min="20" low="50" high="90"></meter>
+<meter max="100" value="10" low="50" optimum="100" high="90"></meter>
+<meter class="ui_meter striped " max="100" value="70" min="0" low="50" high="90"></meter>
+<meter class="ui_meter striped " max="100" value="40" min="20" low="50" high="90"></meter>
+<meter class="ui_meter striped " max="100" value="10" low="50" optimum="100" high="90"></meter>
+
+
+<h3>Meter Fallback</h3>
+<h5>Use ui-meter modifier class for meter</h5>
+
+<div class="ui-meter default">
+	<span  style="width:60%;"></span>
+</div>
+
+<div class="ui-meter warning">
+	<span  style="width:50%;"></span>
+</div>
+
+<div class="ui-meter error">
+	<span  style="width:30%;"></span>
+</div>
+
+<h5>Add  striped to stripe</h5>
+<div class="ui-meter default striped">
+	<span  style="width:60%;"></span>
+</div>
+
+<div class="ui-meter warning striped">
+	<span  style="width:50%;"></span>
+</div>
+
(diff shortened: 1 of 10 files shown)


geekman-rohit:
Its based on #111 , will change and remove the commit if we decide not to use that.

demos/progress.html
```diff
@@ -0,0 +1,125 @@
...
+<body>
+
+<h1>CSS Chassis</h1>
+<h2>Progress and Meter</h2>
+<hr>
+<h3>Meter Element</h3>
+<meter max="100" value="70" min="0" low="50" high="90"></meter>
+<meter max="100" value="40" min="20" low="50" high="90"></meter>
+<meter max="100" value="10" low="50" optimum="100" high="90"></meter>
+<meter class="ui_meter--striped " max="100" value="70" min="0" low="50" high="90"></meter>
```
cvrebert: Why do these 3 use underscores?

demos/progress.html
```diff
@@ -0,0 +1,125 @@
...
+<meter class="ui_meter--striped " max="100" value="40" min="20" low="50" high="90"></meter>
+<meter class="ui_meter--striped " max="100" value="10" low="50" optimum="100" high="90"></meter>
+
+
+<h3>Meter Fallback</h3>
+<h5>Use ui-meter--modifier class for meter</h5>
+
+<div class="ui-meter--default">
```
cvrebert: The `<div>` versions should have ARIA attributes for accessibility; compare with http://getbootstrap.com/components/#progress-basic

cvrebert:
I recommend against the approach of putting the percentage text within the progress bar itself.
(1) It results in inconsistency between `<div>`-based bars and `<progress>`-based bars (one has text, one doesn't)
(2) It breaks at low percentages; example: http://codepen.io/anon/pen/aOMKvZ . There's insufficient room for the text. And even if you tweak the position of the text, it might overlap both the filled and the unfilled portion of the progress bar; it is difficult to ensure that the text color will offer sufficient contrast against the background color in both situations.

As an alternative, I would suggest putting the text directly below the progress bar, and centered horizontally.

scss/molecules/progress/progress/_mixins.scss
```diff
@@ -0,0 +1,55 @@
...
+@mixin ui-progress-fallback-stripes( $fgcolor, $bgcolor ) {
+	@include ui-progress-fallback( $fgcolor, $bgcolor );
+	span {
+		@include ui-stripes();
+	}
+	&--animated {
+		@include ui-progress-fallback( $fgcolor, $bgcolor );
+		span {
+			@include ui-stripes();
+			animation: ui-stripes-animate $ui-progress-animation-time linear infinite;
```
cvrebert: Beware, this leaks memory in Chrome: http://crbug.com/429375

kborchers:
Sorry for jumping in here but it seems like anything around what text and how it's displayed in a progress bar should be left up to a UI lib/framework built on top of Chassis. It feels too overbearing IMO to define that much detail but again, this is just my 2 cents so take it as such.

arschmitz:
@kborchers I agree for the most part though i think we should offer some default styling.

@cvrebert I agree inside is troublesome i think we should opt for very basic styling either above or below what ever we use as a default style for any other label probably ( progress and meter elements are label-able elements ). For the most part unless its a percentage read out the text should probably be a label anyway.

Since we are using BEM that means inheritance is not an issue. That means its fairly trivial to provide default styles that don't matter if they are inside the toolbar or not. Then you can just put a span with the proper class where ever you like to get it properly styled ( and make sure your including proper aria attributes to get it properly associated )

arschmitz:
sorry slipped wrong button DOH!

demos/progress.html
geekman-rohit: sorry they shouldn't have underscores, I kinda forgot to change!

cvrebert:
The pen's `<progress>` doesn't seem to work in Edge:
edge

cvrebert:
[`:indeterminate`](https://developer.mozilla.org/en-US/docs/Web/CSS/:indeterminate) progress bars also need to be considered.

geekman-rohit:
@cvrebert , I need to check how progress works on edge and make changes accordingly, thanks for pointing out!

rohmulan:
Closing for now, until things move on with foms/redoing typography. will open/redo again once other things are done.

PR #128, opened 2015-10-17
Build: Added grunt-html

AndSheCodes2:
Fixes gh-126


package.json
```diff
@@ -43,14 +43,14 @@
...
+    "grunt-contrib-connect": "0.9.0",
     "grunt-contrib-csslint": "0.5.0",
+    "grunt-contrib-cssmin": "0.14.0",
     "grunt-contrib-jshint": "0.11.3",
-    "grunt-contrib-connect": "0.9.0",
     "grunt-contrib-watch": "0.6.1",
     "grunt-csscomb": "3.0.0",
     "grunt-git-authors": "2.0.0",
-    "grunt-html": "1.6.0",
+    "grunt-html": "^5.0.1",
```
jzaefferer: This is interesting, apparently the dependency was already there, but it was unused.

Should remove the `^` to be consistent with other non-ranged dependencies.

jzaefferer:
This is missing the `tasks/options/htmllint.js` file. Not sure how the commit ended up without the new file.

jzaefferer:
@TaMe3971 sorry about the whole rebasing mess, that somehow must ended up with dropping the new file. We can help get that fixed tomorrow.

geekman-rohit:
@sfrisk  @jzaefferer  shouldn't we have removed `^` before landing?

sfrisk:
Yeah, we should have, but I hadn't noticed the `^` when I pulled in the PR. I'll submit a PR to fix it.

PR #150, opened 2016-03-14
RTL : Added CSSjanus to build process

thejdeep:
Added support for RTL using CSSjanus. Fixes Issue #37

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,9 @@
     "grunt-svgstore": "0.5.0",
     "jsass-vars": "0.0.3",
     "load-grunt-config": "0.16.0",
-    "perfjankie": "1.2.2"
+    "perfjankie": "1.2.2",
+    "grunt-cssjanus": "0.2.4",
+    "grunt-concat-css": "0.3.1"
   },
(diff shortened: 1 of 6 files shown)


thejdeep:
@arschmitz Could you please review this sometime ?

sfrisk:
Comments from meeting:

* Need to have RTL and regular CSS in one file so a site could support both options easily
* We should also go through existing scss files to figure out what needs to get flipped and what should be ignored.  In the case where we want things flipped, should we add a demo of these elements?

rawteech:
So you will need have both the regular CSS file and the CSSjanus file as separate mixins but finally compiles to one CSS file I guess

thejdeep:
@sfrisk @arschmitz Please review this sometime. Also, the Travis build fails because the Grunt task crosses 100 chars, can I break down the _build_ task

sfrisk:
After discussing this in today's meeting, we've decided that CSSJanus might not be the best for this particular project, and we should probably be a bit more custom about how we handle RTL users, than generating RTL automatically.

PR #154, opened 2016-03-29
Jsass update

arschmitz:
This pr updates jsass to use grunt-jsass
it also turns on the modules option in jsass and switches to using referencing for variable values


scss/atoms/buttons/_buttons.scss
```diff
@@ -39,49 +39,49 @@
 }
 
 .btn {
-	@include btn(	map-get($ui-btn, margin));
```
geekman-rohit: I thought we were gonna add ui as a prefix in jsass, we are abandoning it completely then ?

arschmitz: i thought we said it didnt make sense to namespace the variables? if we want to we just need to set the namespace option.

sfrisk:
This build is currently failing :-/

PR #161, opened 2016-05-25
New Component: Notifications

nashvail:


sfrisk:
Hey, there was an issue with your signature:

nashvail <nishantve1@gmail.com>: Name on signature (Nishant Verma, Nash Vail) doesn't match name on commit.

scss/atoms/notifications/_mixins.scss
```diff
@@ -0,0 +1,18 @@
...
+
+@mixin notification($type) {
+	border-radius: map-get($notifications-element, border-radius);
+	box-shadow: map-get($notifications-shadows, box-shadow-base);
+	margin: map-get($notifications-element, margin);
+	letter-spacing: map-get($notifications-element, letter-spacing);
+	font-weight: map-get($notifications-element, font-weight);
+	padding: map-get($notifications-element, padding);
+	font-size: map-get($notifications-element, font-size);
+	background: bg($type);
```
sfrisk: You don't need to call functions to get this. Check out how we do it in buttons, just pass in the $notification -success/$notifications-default/etc variable into this mixin and map-get the needed variable from that.  Will make it easier if additional notification colors ever get added, instead of having to add an additional else if statement in both of your function files

sfrisk: See: 
https://github.com/jquery/css-chassis/blob/master/scss/atoms/buttons/_buttons.scss#L45
and https://github.com/jquery/css-chassis/blob/master/scss/atoms/buttons/_mixins.scss#L12

scss/atoms/notifications/_notifications.scss
```diff
@@ -0,0 +1,34 @@
...
+.notification {
+	@include notification(default);
+}
+
+.notification-default {
+	@include notification(default);
+}
+
+.notification-info {
+	@include notification(info);
```
sfrisk: pass in $notifications-info variable here

scss/atoms/notifications/_notifications.scss
```diff
@@ -0,0 +1,34 @@
...
+}
+
+.notification-success {
+	@include notification(success);
```
sfrisk: pass in $notifications-success variable here

scss/atoms/notifications/_notifications.scss
```diff
@@ -0,0 +1,34 @@
...
+}
+
+.notification-warning {
+	@include notification(warning);
```
sfrisk: pass in $notifications-warning variable here

scss/atoms/notifications/_notifications.scss
```diff
@@ -0,0 +1,34 @@
...
+}
+
+.notification-error {
+	@include notification(error);
```
sfrisk: pass in $notifications-error variable here

scss/atoms/notifications/_mixins.scss
sfrisk: so results in:

`background: map-get($type, background);`

scss/atoms/notifications/_mixins.scss
```diff
@@ -0,0 +1,18 @@
...
+	color: text-color($type);
```
sfrisk: replace with:


`background: map-get($type, color);`

scss/variables/notifications.js
```diff
@@ -0,0 +1,70 @@
...
+			// Need help here on setting background color
+			"color": "#ffffff",
+			"background": "#333333"
+		}
+	},
+	"info": {
+		name: "Info Notification",
+		value: {
+			"color": "#ffffff",
+			"background": "#4882B7"
```
geekman-rohit: Use the color palette, add colors to it if you need more, or use shades of it.

geekman-rohit: Fails all Contrast tests for normal size text. Need more contrast in color, but that can wait, and will be fixed automatically if you use palette  colors

scss/variables/notifications.js
```diff
@@ -0,0 +1,70 @@
...
+		}
+	},
+	"element": {
+		name: "Generic Notification Style",
+		value: {
+			"border-radius": "3px",
+			"margin": ".25em",
+			"letter-spacing": "1px",
+			"font-weight": "400",
+			"font-size": "16px",
```
geekman-rohit: em() this

scss/variables/typography.js
```diff
@@ -52,7 +52,7 @@ chassis.typography = {
 	},
 	"sans": {
 		name: "Font Family - Sans",
-		value: "Helvetica, Arial, sans-serif"
+		value: "Open Sans, Helvetica, Arial, sans-serif"
```
geekman-rohit: Open Sans can be arguably better than helvetica, but something that should be discussed in meeting

demos/notifications/notifications.html
```diff
@@ -0,0 +1,19 @@
...
+	<title>CSS Chassis - Typography</title>
+	<meta name="description" content="Typography skeleton for styling">
+	<meta name="viewport" content="width=device-width, initial-scale=1">
+	<link rel="stylesheet" href="../../dist/css/chassis.css">
+	<link rel="stylesheet" href="notifications.css">
+	<link href="http://fonts.googleapis.com/css?family=Open+Sans:400,600,700,400italic,700italic" rel="stylesheet">
+</head>
+<body>
+	<div class="container">
+		<div class="notification notification-default">
```
geekman-rohit: More examples?

sfrisk: Should have a demo for each option you've created

scss/atoms/notifications/_notifications.scss
```diff
@@ -0,0 +1,34 @@
...
+@import
+	"dist/chassis",
+	"functions",
+	"mixins";
+
+.notification {
+	@include notification(default);
+}
+
+.notification-default {
```
sfrisk: You can simplify writing this by using `&-default` underneath the `.notification` class.  Not sure if we want to organize things that way, but it would prevent having to type "notification" over and over again.

scss/variables/notifications.js
nashvail: You mean to wrap px values in em? Like this em("3px") ?

nashvail: Gives a em() not defined error on doing em("3px")

demos/notifications/cayman.css
```diff
@@ -0,0 +1,241 @@
+* {
```
geekman-rohit: Whether we should use some external css theme for demo page at this stage is debatable. Either ways the file is external and should go in external directory @arschmitz thoughts?

nashvail: I have opened an issue for discussion on the same, #162

demos/notifications/normalize.css
```diff
@@ -0,0 +1,424 @@
+/*! normalize.css v3.0.2 | MIT License | git.io/normalize */
+
+/**
+ * 1. Set default font family to sans-serif.
+ * 2. Prevent iOS text size adjust after orientation change, without disabling
+ *    user zoom.
```
geekman-rohit: We already have normalize in external

scss/variables/colors.js
```diff
@@ -24,7 +25,8 @@ chassis.colors = {
 			"base": "#3F51B5",
 			"light": "#5C6BC0",
 			"dark": "#3949AB",
-			"darker": "#303F9F"
+			"darker": "#303F9F",
+			"notification": "#3F51B5"
```
geekman-rohit: This is the base color. Why add one more key?

scss/variables/colors.js
```diff
@@ -33,7 +35,8 @@ chassis.colors = {
 			"base": "#43A047",
 			"light": "#4CAF50",
 			"dark": "#388E3C",
-			"darker": "#2E7D32"
+			"darker": "#2E7D32",
+			"notification": "#43A047"
```
geekman-rohit: check comment above

scss/variables/notifications.js
```diff
@@ -0,0 +1,68 @@
...
+			"padding": "15px 15px 15px 25px"
+		}
+	},
+	"default": {
+		name: "Default Notification",
+		value: {
+			"color": "#ffffff",
+			"background": () => "colors.default"
```
geekman-rohit: :+1:

scss/variables/colors.js
nashvail: The "default" color's base and default notification's color differ hence I decided to add a notification key in each of the color instead of just adding in one and creating a special case of default in mixin.scss

scss/variables/notifications.js
geekman-rohit: yes

scss/variables/notifications.js
geekman-rohit: background: ()=> "colors.default.base" maybe? I think jsass-vars supports fetching a map value upto one level.  Will check and revert anyways. @arschmitz  ?

scss/atoms/notifications/_mixins.scss
```diff
@@ -0,0 +1,21 @@
...
+	border-radius: map-get($notifications-element, border-radius);
+	margin: map-get($notifications-element, margin);
+	letter-spacing: map-get($notifications-element, letter-spacing);
+	font-weight: map-get($notifications-element, font-weight);
+	padding: map-get($notifications-element, padding);
+	font-size: map-get($notifications-element, font-size);
+	box-shadow: map-get($notifications-shadows, box-shadow-base);
+
+	// Type specific styles
+	background: map-deep-get($type, background, notification);
```
geekman-rohit: Makes changing the variable value very difficult. the $type.background should ideally represent the color code itself, fetched from the palette

scss/variables/colors.js
geekman-rohit: You can add a default.grey or a new palette 'grey':{'dark':#333333} Lets discuss in next meeting.

 Anyways if you set your background  variable directly to the color code value rather than the entire map, things will get simpler and allow you more choice, since then you need not access using map-deep-get(, , notifications) everytime.

scss/variables/notifications.js
geekman-rohit: em() is an scss function should be in your scss

scss/atoms/notifications/_mixins.scss
```diff
@@ -0,0 +1,18 @@
...
+*/
+
+@mixin notification($type) {
+	border-radius: map-get($notifications-element, border-radius);
+	box-shadow: map-get($notifications-shadows, box-shadow-base);
+	margin: map-get($notifications-element, margin);
+	letter-spacing: map-get($notifications-element, letter-spacing);
+	font-weight: map-get($notifications-element, font-weight);
+	padding: map-get($notifications-element, padding);
+	font-size: map-get($notifications-element, font-size);
```
geekman-rohit: em( map-get($notifications-element, font-size))

demos/notifications/cayman.css
sfrisk: yeah, an external library shouldn't be included here.

demos/notifications/normalize.css
sfrisk: and compiled into chassis.css

demos/notifications/notifications.html
```diff
@@ -0,0 +1,69 @@
...
+</head>
+<body>
+  <section class="main-content">
+		<h1>Notifications</h1>
+		<hr>	
+  	<!-- Default Notification -->
+		<div class="notification notification-default">
+			<strong>Title!</strong> Alert body...
+		</div>			
+<pre><code>&lt;div class="notification notification-default"&gt;
```
sfrisk: There are JS libraries we could use to do this for us, as I mentioned in the meeting, instead of having to type everything twice and hoping we typed it the same way each time.

scss/variables/colors.js
sfrisk: I agree with @geekman-rohit, I don't think we should be adding another variable like this indicating the notification color, especially if it's often the same as the base.

scss/variables/typography.js
sfrisk: Also a font that would have to be included.

Also in general this sort of typography change probably shouldn't be occurring within the scope of this PR - but something that should be brought up in a larger discussion.

demos/notifications/notifications.html
sfrisk: (So for now we should probably remove the demo versions of these, come up with a separate PR to figure out how we want to handle demo code cross project)

scss/atoms/notifications/_functions.scss
```diff
@@ -0,0 +1,6 @@
+/*
```
sfrisk: If you aren't doing anything with this file, remove it

scss/atoms/notifications/_notifications.scss
```diff
@@ -0,0 +1,37 @@
...
+
+.notification-warning {
+	@include notification-colors($notifications-warning);
+}
+
+.notification-error {
+	@include notification-colors($notifications-error);
+}
+
+.notification .notification-title {
```
kristyjy: This should probably be just .notification-title instead of .notification .notification-title to avoid cascading.

PR #163, opened 2016-05-27
New Component: Modals

nashvail:


cvrebert:
I suggest including an example where the modal's content is taller than the viewport (and thus, scrolling is required).

scss/atoms/modals/_modals.scss
```diff
@@ -0,0 +1,33 @@
...
+	"dist/chassis",
+	"functions",
+	"mixins";
+
+.modal {
+	@include modal();
+
+	.modal-header {
+		@include modal-header();
+		h4 {
```
cvrebert: Hardcoded tag selector :-/

scss/atoms/modals/_modals.scss
```diff
@@ -0,0 +1,33 @@
...
+	}
+
+	.modal-body {
+		@include modal-body();
+	}
+
+	.modal-footer {
+		@include modal-footer();
+		@include clearfix();
+		& .btn {
```
cvrebert: Probably too opinionated to do this by default.

scss/atoms/modals/_modals.scss
nashvail: The default one as styled in typography.scss has wide top and bottom margins so I had to override that.

cvrebert: What about a `.modal-title` class then?

sfrisk: In general we want to avoid cascading as well. Have modal-title so they can use whatever element they need semantically, and remove modal-header cascading from inside modal.

scss/atoms/modals/_modals.scss
geekman-rohit: And that is very very weird?? 
```sass
.modal-footer {
  &  .btn  {
  }
}
```
whats wrong with
```sass
.modal-footer {
  .btn  {
  }
}
```
And either ways, I would rather have a separate class .modal-button .
And what @cvrebert  said, floating to right directly is very opinionated, that should be a variable as well

something like
```sass
.modal-button {
    float: map-get(modal-button,float)
}
```

scss/atoms/modals/_mixins.scss
```diff
@@ -0,0 +1,47 @@
+/*
+* ==========================================================================
+* File: _mixins.scss
+* For : Modals
+* ==========================================================================
+*/
+
+// Styles shared by modal's inner components
+$border-radius: map-get($modal-borders, radius);
+$padding: em(map-get($modal-padding, padding));
```
nashvail: Retrieved externally because apparently scss was compiling "0 15px" to just "15px". Let me know if I am doing something wrong.

sfrisk:
button stacking is a little awkward in mobile.

demos/modals/modals.css
```diff
@@ -0,0 +1,5 @@
+body {
```
sfrisk: for something like this, maybe a common demo.css file would be better, than separate stylesheets for everything?

scss/atoms/modals/_modals.scss
```diff
@@ -0,0 +1,40 @@
...
+.modal {
+	@include modal();
+}
+
+.modal-header {
+	@include modal-header();
+}
+
+.modal-title {
+	margin: 0;
```
kristyjy: You shouldn't need margin: 0 here because the mixin appears to also include a margin declaration.

scss/variables/modal.js
```diff
@@ -0,0 +1,75 @@
...
+			"margin": "10px",
+			"background": chassis.colors.background.value.base
+		}
+	},
+	"title": {
+		name: "Modal Title Styles",
+		value: {
+			"font-size": "24px",
+			"font-weight": "700",
+			"margin": "0"
```
kristyjy: We should be consistent about how we write 0: 0 vs 0px (the rest of this file seems to be using "0px") I prefer using 0 myself but we should talk about it in our next meeting maybe?

sfrisk: CSS style guide for jQuery states to not use units for 0s. https://contribute.jquery.org/style-guide/css/

nashvail: Executing em function (https://github.com/jquery/css-chassis/blob/master/scss/atoms/typography/_functions.scss) on 0 values without a unit results in a warning "Warning: 0em/px isn't a valid CSS value."

Issue #165, opened 2016-05-31
Button styling when used on links shouldn't underline on hover


sfrisk:
Currently when hovering over links styled like buttons, and underline appears.  Make sure this doesn't occur.

PR #166, opened 2016-06-01
Buttons: Removed underline from link buttons on hover

nashvail:
For issue #165

### DIFF
--- a/scss/atoms/buttons/_mixins.scss
+++ b/scss/atoms/buttons/_mixins.scss
@@ -36,6 +36,10 @@
 		@include btn-hover($btn-colors);
 	}
 
+	&:hover {
+		text-decoration: none;
+	}
+
 	&:disabled,
 	&.disabled {
 		@include btn-disabled($btn-colors);


scss/atoms/buttons/_buttons.scss
```diff
@@ -38,6 +38,12 @@
 	width: 100%;
 }
 
+// To remove the underline that appears when hovering link styled buttons
+a.btn:hover,
+a.btn.hover {
+	text-decoration: none;
```
sfrisk: I wouldn't add this here, but as part of the hover related mixin.

scss/atoms/buttons/_mixins.scss
```diff
@@ -36,6 +36,10 @@
 		@include btn-hover($btn-colors);
 	}
 
+	&:hover {
+		text-decoration: none;
+	}
```
geekman-rohit: why not just add it to :hover/.hover above ??

sfrisk: .hover already inherits the text-decoration: none from the btn mixin up above (line 8).  This is dealing with browser defaults overwriting the pseudo class functionality.

PR #174, opened 2017-01-12
Typography: change the unit measurement from 'em' to 'rem'

lunaticmonk:
I tried to refactor the code and changed the em converter to rem since it would be better to use rem. em would cause problems when 2-3 elements are nested. Opinions?

### DIFF
--- a/scss/atoms/typography/_functions.scss
+++ b/scss/atoms/typography/_functions.scss
@@ -11,6 +11,6 @@
 * 1. font-size: em(14px);
 * 2. font-size: em(30px/14px);
 */
-@function em($value, $context: map-get($typography-font-default, font-size)) {
-	@return ($value / $context * 1em);
+@function rem($value, $context: map-get($typography-font-default, font-size)) {
+	@return ($value / $context * 1rem);
 }
--- a/scss/atoms/typography/_typography.scss
+++ b/scss/atoms/typography/_typography.scss
@@ -110,7 +110,7 @@ strong {
 
 code {
 	background: map-get($typography-code,background);
-	font-size: em( 16px );
-	padding: em( 4px, 16px ) em( 8px, 16px );
+	font-size: rem( 16px );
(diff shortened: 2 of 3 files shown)


rohmulan:
@sumedh123 just changing the unit will not suffice in our case. It's not so much change the unit that we wish to do, but change the approach entirely. This will most probably mean a major code change and a change/removal of the em mixin we are using

scss/atoms/typography/_functions.scss
```diff
@@ -11,6 +11,6 @@
 * 1. font-size: em(14px);
 * 2. font-size: em(30px/14px);
 */
-@function em($value, $context: map-get($typography-font-default, font-size)) {
-	@return ($value / $context * 1em);
+@function rem($value, $context: map-get($typography-font-default, font-size)) {
+	@return ($value / $context * 1rem);
```
rohmulan, line 15: We should reconsider this approach of deciding in pixels and then converting to ems/rems. It should be the other way round, deciding a scale then converting it to pixels(only if necessary). I'll bring it up in next meeting. We can keep the PR pending until then.

scss/atoms/typography/_typography.scss
```diff
@@ -110,7 +110,7 @@ strong {
 
 code {
 	background: map-get($typography-code,background);
-	font-size: em( 16px );
-	padding: em( 4px, 16px ) em( 8px, 16px );
+	font-size: rem( 16px );
+	padding: rem( 4px, 16px ) rem( 8px, 16px );
```
rohmulan, line 114: Padding should probably be em not rem. Again a point we can discuss during next meeting. Padding is meant to be space between border and text to make text look better and readable, it makes sense to decide padding in scale of current element font size rather than root element font size.

rohmulan (changes requested):

rohmulan:
The changes I requested all need discussion, but they need to be discussed before we merge any more PRs

